PR #1, opened 2020-11-20
Fix gitorious and dfu-util links

tormodvolden:
Replace gitorious.org links (site no longer existing) and update dfu-util links (host changed many years ago).

Also remove the search index which isn't maintained by hand.


Pillar1989:
merged, thanks

PR #94, opened 2023-03-23
Add:  change the picture size for SenseCAP S2110 Sensor Builder "

Extremelymeng:
Hello, this is Yvonne!
I am editing SenseCAP S2110 Sensor Builder on the wiki contents, please help me merge it.


MatthewJeffson (approved):
Change every wrong width "" which is kind of OK

PR #104, opened 2023-03-24
Get one now Changed

KasunThushara:
The button "Get one now" changed to a new design on OS installation page


MatthewJeffson (approved):
thank you for your conribution!

PR #107, opened 2023-03-27
FIX: reterminal using key with tp_rotate=1 error

peterpanstechland:


MatthewJeffson (approved):
thank your for contribution

PR #123, opened 2023-04-04
update fw link and reg table for sensor builder

Maxwelltoo:


MatthewJeffson (approved):
Thank you for your adding up!

PR #133, opened 2023-04-10
Fix:xiao esp32s3 pin content

limengdu:


MatthewJeffson (approved):
thank you for the add up!

PR #161, opened 2023-04-19
Update: A607 Orin Nano flash and USB flash

lakshanthad:


MatthewJeffson (approved):
Thank you for the update!

PR #163, opened 2023-04-20
Update: YOLOv8 wiki with multistream config and benchmarks

lakshanthad:


MatthewJeffson (approved):
Thank you for the update

PR #196, opened 2023-05-08
Update XIAO ESP32C3 IO allocation

matsujirushi:


MatthewJeffson (approved):
Thank you for the fix

PR #210, opened 2023-05-12
Add: XIAO nRF52840 using QSPI Flash.

limengdu:


MatthewJeffson (approved):
Thank you for the add up

PR #211, opened 2023-05-12
Fix: Orin NX 16GB flashing command

lakshanthad:


MatthewJeffson (approved):
thank you for the fix!

PR #223, opened 2023-05-18
reTerminal and Pi camera module 

KasunThushara:
Grammer mistakes were corrected review 01


MatthewJeffson (approved):
Thank you for the update

PR #225, opened 2023-05-18
reTerminal Picam data correction 

KasunThushara:
Data correction and ad a note


MatthewJeffson (approved):
Thank you for the update

PR #227, opened 2023-05-19
Add: xiao esp32s3 uf2 bootloader

limengdu:


MatthewJeffson (approved):
Thank you for your add up

PR #243, opened 2023-05-25
Update: SenseCAP Indicator Flash Tutorial and fullfill API key

Love4yzp:


MatthewJeffson (approved):
Thank you for the add up!

PR #250, opened 2023-05-26
add links plex and timezone

KasunThushara:
add links plex and timezone


MatthewJeffson (approved):
Thank you for the update!

PR #251, opened 2023-05-27
Revise project description in  Grove_System.md

Pigpog:
Reduce the verbosity of the first two paragraphs.

These shorter paragraphs introduce the project in fewer sentences, retaining key ideas.


MatthewJeffson (approved):
Thank you for your update! Some of these wikis have been around for a long time, and I am sure your updates will bring good results.

PR #254, opened 2023-05-29
SenseCAP Indicator Flash firmware wiki

Love4yzp:


MatthewJeffson (approved):
Thanks for the update!

PR #267, opened 2023-06-02
modify to Train-Water-Meter-Digits-Recognition-Model-with-SenseCAP-A1101

Kway99:
Modify the file name to Train-Water-Meter-Digits-Recognition-Model-with-SenseCAP-A1101. Thanks for checking~


MatthewJeffson (approved):
Thank you for the update

PR #274, opened 2023-06-05
Add: description of situations where millimetre wave radar may cause anomalies

limengdu:


MatthewJeffson (approved):
Thanks for the update

PR #280, opened 2023-06-06
Update: SenseCAP Indicator HA

Love4yzp:


MatthewJeffson (approved):
thanks for the update

PR #281, opened 2023-06-06
Add: XIAO CAN Bus expansion board wiki

limengdu:


MatthewJeffson (approved):
Thanks for the update

PR #288, opened 2023-06-08
Fix: Grove SGP40 picture

limengdu:


MatthewJeffson (approved):
Thanks for the update

PR #293, opened 2023-06-09
"UPDATE: lora module usage for reterminal dm and fix link type in reterminalbridge"

peterpanstechland:


MatthewJeffson (approved):
Thanks for the update!

PR #313, opened 2023-06-13
FIX: reterminal dm getting start with ned red

peterpanstechland:


MatthewJeffson (approved):
Thanks for the update

PR #333, opened 2023-06-14
Update Plex-Linkstar with NAS

KasunThushara:
Mount Plex Directory on the PC and Transfer the Media chapter was added


MatthewJeffson (approved):
Thanks for the update!

PR #355, opened 2023-06-27
Fix: xiao nrf52840 sd card usage

limengdu:


MatthewJeffson (approved):
Thanks for the add up

PR #356, opened 2023-06-27
ADD: EdgeBox RPi 200 boot From External Media

peterpanstechland:


MatthewJeffson (approved):
Thanks for the add up

PR #359, opened 2023-06-28
Fix: Delete fan usage and add system monitor

lakshanthad:


MatthewJeffson (approved):
Thanks for the update

PR #363, opened 2023-06-29
Update: Add flash preconfigure

lakshanthad:


MatthewJeffson (approved):
Thanks for the update

PR #364, opened 2023-06-29
Update: Add Orin Nano flashing

lakshanthad:


MatthewJeffson (approved):
Thanks for the update

PR #367, opened 2023-06-29
Fix: Fix desc of XIAO ESP32S3 pin multiplexing doc

mobyw:


MatthewJeffson (approved):
Thank you for the fix!

PR #368, opened 2023-06-30
UPDATE: Edgebox-RPI-200 boot from external media

peterpanstechland:


MatthewJeffson (approved):
Thanks for the update

PR #374, opened 2023-07-04
Fix: xiao esp32c3 I2S content

limengdu:


MatthewJeffson (approved):
Thanks for the update

PR #383, opened 2023-07-06
Update: XIAO/ODYSSEY/Jetson/reTerminal FAQ page

grilliiiii:


MatthewJeffson (approved):
Thanks for the update

PR #391, opened 2023-07-10
Update Grove-PIR_Motion_Sensor documentation

RitchieP:
Update a little bit of documentation which I think might be helpful for some beginners when learning about the PIR Motion Sensor.


MatthewJeffson (approved):
Wow! Thanks for the update and the knowledge shared!

PR #398, opened 2023-07-14
add JFlash part and change some settings

Jessie219-web:
dd JFlash part and change some settings


MatthewJeffson (approved):
Thank you for the update!

PR #405, opened 2023-07-18
modify the pins

Jessie219-web:
modify the pin connections and server configuration


MatthewJeffson (approved):
Thank you for the update!

PR #407, opened 2023-07-19
Update: 4G module usage on reterminal DM 

peterpanstechland:


MatthewJeffson (approved):
Thank you for the update

PR #408, opened 2023-07-19
FIX: Flash OS confusion with 32bit driver and update sensecraft edge OS download link

peterpanstechland:


MatthewJeffson (approved):
Thank you for the update

PR #411, opened 2023-07-20
ADD: Edgebox-rpi-200 setup balena OS

peterpanstechland:


MatthewJeffson (approved):
Thanks for the update

PR #416, opened 2023-07-21
add libraries installation steps

Jessie219-web:
add libraries installation steps on node-red configuration


MatthewJeffson (approved):
Thanks for the update

PR #417, opened 2023-07-21
Update: (XIAO Canbus Expansion board)Change the pin prompt for the ha…

grilliiiii:
…rdware overview


MatthewJeffson (approved):
Thanks for the update

PR #445, opened 2023-07-25
Add: XIAO ESP32S3 Edge Impulse Wiki

limengdu:


MatthewJeffson (approved):
Thanks for the add up

PR #452, opened 2023-07-26
delete some old links

Jessie219-web:
delete some broken links


MatthewJeffson (approved):
Thanks for the update

PR #462, opened 2023-07-28
add OpenWrt knowledge database

Yuluer:
add OpenWrt knowledge database


MatthewJeffson (approved):
Thanks for the update

PR #469, opened 2023-07-31
Update: add the note on linkit

grilliiiii:


MatthewJeffson (approved):
Thanks for the update

PR #486, opened 2023-08-08
FIX: balena lowercase typo

peterpanstechland:


MatthewJeffson (approved):
Thanks for the update

PR #496, opened 2023-08-11
modify the wrong code

Jessie219-web:
modify the wrong code, and add the demo effect picture


MatthewJeffson (approved):
Thanks for the update!

PR #497, opened 2023-08-12
added about EI and organised the page for tinyML workshop. 

salmanfarisvp:
Just added about EI and now started Organising for the tinyML workshop docs.


MatthewJeffson (approved):
Thanks for the add up!

PR #508, opened 2023-08-15
add tracker get start

Jessie219-web:
add get start with SenseCAP T1000 Tracker


MatthewJeffson (approved):
Thanks for the add up!

PR #525, opened 2023-08-16
add tracker Helium services

Jessie219-web:
add SenseCAP tracker Helium services


MatthewJeffson (approved):
Thanks for the add up!

PR #536, opened 2023-08-18
Fix git branch name in contribution guide

aloysiousBenoy:
Changed the default branch name in the sample git command to reflect what we are using right now. The branch ```main``` is not present in the repository and we are using ```docusaurus-version``` as the default branch name


MatthewJeffson (approved):
Oh, thank you for pointing it out!

PR #540, opened 2023-08-21
correct misspelled words for T1000 tracker

Jessie219-web:
correct some misspelled words for SenseCAP T1000 tracker


MatthewJeffson (approved):
Thanks for the update!

PR #542, opened 2023-08-21
reTerminal with HA and Frigate intergration

KasunThushara:
reTerminal with HA and Frigate intergration


MatthewJeffson (approved):
Thank you for the add up!

PR #543, opened 2023-08-23
change the KS status on SenseCAP T1000 Tracker

Jessie219-web:
change the KS status to LIVE on SenseCAP T1000 Tracker


MatthewJeffson (approved):
thanks for the update!

PR #544, opened 2023-08-23
Add SenseCAP LoRaWAN Starter Kit

Kway99:
I have added sensecap lorawan starter kit. Please kindly have check! thanks for your time~


MatthewJeffson (approved):
Thanks for the update

PR #545, opened 2023-08-24
add SenseCAP decoder

Jessie219-web:
add SenseCAP decoder for LoRaWAN devices


MatthewJeffson (approved):
Thanks for the add up!

PR #549, opened 2023-08-25
syntax correction and adding wifi troubleshoot

KasunThushara:
syntax correction and add wifi troubleshoot after installation.


MatthewJeffson (approved):
Thanks for the update!

PR #559, opened 2023-08-28
modified version of HA and reTerminal

KasunThushara:
Remove portainer and add some troubleshooting cases under notes.


MatthewJeffson (approved):
Thanks for the update!

PR #560, opened 2023-08-28
add M2 picture

Jessie219-web:
update pictures


MatthewJeffson (approved):
Thanks for the update

PR #561, opened 2023-08-29
update AWS of M2

Jessie219-web:
update AWS of SenseCAP M2 Multi platform


MatthewJeffson (approved):
Thanks for the update

PR #563, opened 2023-08-29
Update: mmwave kit url

limengdu:


MatthewJeffson (approved):
Thanks for the update

PR #564, opened 2023-08-29
delete the wrong url for M2

Jessie219-web:
delete the wrong url


MatthewJeffson (approved):
Thanks for the updating!

PR #566, opened 2023-08-30
Fix: AS3935 code bug

limengdu:


MatthewJeffson (approved):
Thanks for the update

PR #567, opened 2023-08-30
add M2 LNS Configuration

Jessie219-web:
add LNS Configuration for SenseCAP M2 gateway


MatthewJeffson (approved):
Thanks for the add up!

PR #574, opened 2023-09-04
add AWS for T1000 tracker

Jessie219-web:
add AWS for T1000 tracker


MatthewJeffson (approved):
Thanks for the update!

PR #577, opened 2023-09-05
Update: xiao ble sidewalk

limengdu:


MatthewJeffson (approved):
Thanks for the update

PR #586, opened 2023-09-12
Update: Change one-line script and reorganize

lakshanthad:


MatthewJeffson (approved):
Thanks for the update

PR #587, opened 2023-09-12
Reterminal and edgeimpulse

KasunThushara:
Reterminal and edgeimpulse version 1


MatthewJeffson (approved):
Thanks for the update!

PR #593, opened 2023-09-13
update name for wio tracker 1110

Jessie219-web:


MatthewJeffson (approved):
Thanks for the update

PR #600, opened 2023-09-15
add home assistant and sensecap lorawan sensor

Jessie219-web:


MatthewJeffson (approved):
Thanks for the update!

PR #601, opened 2023-09-15
add applications for sensecap sensor

Jessie219-web:


MatthewJeffson (approved):
Thanks for the update!

PR #606, opened 2023-09-18
update old link for sensecap decoder

Jessie219-web:


MatthewJeffson (approved):
Thanks for the update

PR #607, opened 2023-09-19
update SenseCAP Hotspot APP and bazaar links for hotspot

Jessie219-web:


MatthewJeffson (approved):
Thanks for the update and the add up

PR #608, opened 2023-09-19
update wio tracker 1110 DB

Jessie219-web:


MatthewJeffson (approved):
Thanks for the update

PR #609, opened 2023-09-19
Add: xiao io expander board

limengdu:


MatthewJeffson (approved):
Thanks for the update

PR #610, opened 2023-09-20
add new section on edgetpu

KasunThushara:
add new section on edgetpu


MatthewJeffson (approved):
Thanks for the update

PR #613, opened 2023-09-21
Add: FAQ for SenseCAP Tracker

Jessie219-web:


MatthewJeffson (approved):
Thanks for the add up and the js function updated

PR #614, opened 2023-09-21
edge impulse and reterminal DM

KasunThushara:
edge impulse and reterminal DM


MatthewJeffson (approved):
Thanks for the add up!

PR #645, opened 2023-10-07
Add: xiao esp32s3 sense ov5640 camera datasheet

limengdu:


MatthewJeffson (approved):
Thanks for the update

PR #656, opened 2023-10-10
Add: add L76-L GNSS for XIAO wiki

limengdu:


MatthewJeffson (approved):
Thanks for the add up

PR #660, opened 2023-10-11
Edge Impulse for Keyword Spoting using XIAO nRF52840 Sense

feiticeir0:
This is a tutorial on how to use Edge Impulse with the XIAO nRF52840 to run inference on ML models using the PDM microphone.


MatthewJeffson (approved):
Hello! I have updated some contents here.
Thank you for doing it!:D

PR #664, opened 2023-10-13
Change Shipping OS to Raspberry Pi

KasunThushara:
Change Shipping OS to Raspberry Pi


MatthewJeffson (approved):
Thanks for the update

PR #667, opened 2023-10-16
Update: YOLOv8 benchmarks with single and multi models

lakshanthad:


MatthewJeffson (approved):
Thanks for the update

PR #677, opened 2023-10-18
Add: modify RG-15 and RG-9 file

Extremelymeng:
Hello, this is Yvonne! I am adding the RG-15 and RG-9 files to the wiki contents, please help me merge them.


MatthewJeffson (approved):
Thanks for the update and the add up!

PR #682, opened 2023-10-19
Add: xiao gpio board connection way

limengdu:


MatthewJeffson (approved):
Thanks for the update

PR #706, opened 2023-10-23
Update SenseCAP_M2_Data_Only_Quick_Start.md with fees info

waveform06:
Updated with better info on the onboarding and location assert fees


MatthewJeffson (approved):
Thanks for the update!

PR #727, opened 2023-10-27
Add: HACS Install

QingWind6:


MatthewJeffson (approved):
Thanks for the update!

PR #772, opened 2023-11-03
FIX: XIAO eInk Expansion board fpc numbers

peterpanstechland:


MatthewJeffson (approved):
Thanks for the update!

PR #775, opened 2023-11-06
FIX: FPC pin number error

peterpanstechland:


MatthewJeffson (approved):
Thanks for the update

PR #781, opened 2023-11-07
Update: add CN version of XIAO SAMD21 and nRF52840-Sense

zgy20020114:


MatthewJeffson (approved):
Thank you for the add up!

PR #782, opened 2023-11-08
picam and opencv with reterminal tutorial

KasunThushara:
picam and opencv with reterminal tutorial


MatthewJeffson (approved):
Thanks for the update!

PR #786, opened 2023-11-09
Update:Add CN version of XIAO nRF52840-Sense-Embedded ML

JessieJJ5:


MatthewJeffson (approved):
Thanks for the add up!

PR #792, opened 2023-11-16
Update installation guide

KasunThushara:
add a bookwormOS image path in black screen issue page. add some notes about bookworm 64 bit and Bullseye 32 and 64-bit.


MatthewJeffson (approved):
Thanks for the update!

PR #793, opened 2023-11-16
Update URLs opencv and add computer vision to reterminal collection page 

KasunThushara:
Fixed broken URLs opencv and add computer vision to reterminal collection page


MatthewJeffson (approved):
Thanks for the update!

PR #809, opened 2023-11-24
Update:add Incorrect screen orientation on RPiOS Bookworm method

grilliiiii:


MatthewJeffson (approved):
Thanks for the update!

PR #814, opened 2023-11-28
add new application for jetson 

yuyoujiang:
The name of application is `reComputer is all you need`


MatthewJeffson (approved):
Thank you for the add up!

PR #821, opened 2023-12-04
add wio tracker dev kit for sidewalk

Jessie219-web:


MatthewJeffson (approved):
Thanks for the update and the add up!

PR #826, opened 2023-12-06
Fix:change wiki title

yuyoujiang:
from `reComputer is all you need` to How to train and deploy YOLOv8 on reComputer


MatthewJeffson (approved):
Thanks for the update

PR #827, opened 2023-12-07
update: XIAORP2040 update a note

Allenkzl:


MatthewJeffson (approved):
Thanks for the update

PR #829, opened 2023-12-07
Update wiki with bookworm objectdetection

KasunThushara:
Update wiki with bookworm objectdetection


MatthewJeffson (approved):
Thanks for the update!

PR #836, opened 2023-12-11
Fix character code for space

matsujirushi:
image


MatthewJeffson (approved):
thank you for the fix!

PR #841, opened 2023-12-12
Modified：Changing details for reSpeaker 4 Mic Linear and Lora E5

grilliiiii:


MatthewJeffson (approved):
Thanks for the update

PR #843, opened 2023-12-12
Object tracking opencv pi cam

KasunThushara:


MatthewJeffson (approved):
Thank you for the add up

PR #844, opened 2023-12-12
add: mmwave kit category add MMWave Kit And Grove Connect To ESPHome

Allenkzl:


MatthewJeffson (approved):
Thank you for the update!

PR #845, opened 2023-12-13
Append default boot.conf of reTerminal

matsujirushi:
* Links:
   * https://lab.seeed.co.jp/entry/2022/01/05/120000


MatthewJeffson (approved):
Thank you for the update!

PR #861, opened 2023-12-20
fixed slugs for better SEO

KasunThushara:
fixed slugs for better SEO


MatthewJeffson (approved):
Thanks for the update

PR #867, opened 2023-12-22
add dev kit flash guide

Jessie219-web:


MatthewJeffson (approved):
thanks for the update

PR #871, opened 2023-12-24
Update reterminal and DM collection pages, update oddessy 

KasunThushara:


MatthewJeffson (approved):
Thanks for the update

PR #884, opened 2024-01-02
Add: a new demo of LLM

yuyoujiang:
How to Format the Output of LLM Using Langchain on Jetson


grilliiiii (approved):
Thanks for the update

PR #900, opened 2024-01-10
Add: Use Infineon Wifi Module on reComputer Jetson

yuyoujiang:
Add: Use Infineon Wifi Module on reComputer Jetson
Add: Solution for the compatibility issue between Seeed carrier board J401 and VEYE camera


grilliiiii (approved):
Thanks for your updates！

PR #914, opened 2024-01-14
Add: local voice chatbot

yuyoujiang:
- add a new demo named local voice chatbot
- fix the problem about `reComputer_Jetson_with_Infineon_Wi-Fi_Module.md`


MatthewJeffson (approved):
Thanks for the add up!

PR #920, opened 2024-01-16
Update: vision ai v2 update some files

Allenkzl:


MatthewJeffson (approved):
Thanks for the update

PR #930, opened 2024-01-22
Fix: grove vision ai connect to mate app wiki

limengdu:


grilliiiii (approved):
thanks for your updates!

PR #932, opened 2024-01-22
update: xiao ble update a tip about serial function can't use in ada develpment board.

Allenkzl:
…develpment board.


MatthewJeffson (approved):
Thanks for the update

PR #934, opened 2024-01-22
update: vision ai v2 add BOM

Allenkzl:


MatthewJeffson (approved):
Thanks for the update

PR #947, opened 2024-01-29
update: sunlight sensor update SI1145 example and fix Si1151 example

Allenkzl:


grilliiiii (approved):
thanks for your update!

PR #951, opened 2024-01-30
delete: detele ardupy with wio terminal

Allenkzl:


grilliiiii (approved):
thanks for your updates!

PR #953, opened 2024-01-31
update: sunlight sensor update a link

Allenkzl:


grilliiiii (approved):
Thanks for your updates!

PR #956, opened 2024-01-31
update: mp3 V4 module update some info

Allenkzl:


grilliiiii (approved):
Thanks for the updates!

PR #958, opened 2024-02-02
update: LCD RGB Backlight model update MBED support

Allenkzl:


grilliiiii (approved):
Thanks for the updates!

PR #963, opened 2024-02-05
ADD: Shape Detection with OpenCV wiki 

KasunThushara:
ADD: Shape Detection with OpenCV wiki


MatthewJeffson (approved):
Thanks for the add up!

PR #974, opened 2024-02-21
updata: vision v2 update a tip about camera

Allenkzl:


MatthewJeffson (approved):
Thank you for the update

PR #982, opened 2024-02-26
fix: delete a tip about camera support

Allenkzl:


MatthewJeffson (approved):
Thanks for the update!

PR #992, opened 2024-02-28
update: add xiao 2040 and SAMD21 to this topic

Allenkzl:


grilliiiii (approved):
Thanks for your updates!

PR #993, opened 2024-02-28
Add Raspberry Pi Pico instructions to Gear Stepper Motor Driver Pack

Binney:
Tried these on my Pi Pico and it works. Also added a bit of context for what exactly a stepper motor is and how this makes it work.

I have read and agree to the individual CLA


MatthewJeffson (approved):
That is really nice of you! Thanks for doing it!

PR #994, opened 2024-02-28
Gear Stepper Motor Driver Pack: Remove commented out code

Binney:
Not sure why this was here but it makes it a lot harder to read and novices may not understand what a comment is


MatthewJeffson (approved):
Yeah, that is the old time wiki content issues. We really need the help like yours.

PR #999, opened 2024-03-01
Update: add ubuntu 22.04

yuyoujiang:
add gpio table of ubuntu22.04 for ODYSSEY


grilliiiii (approved):
Thanks for your updates!

PR #1007, opened 2024-03-05
fix: fix grove vision ai v2 CH343 driver download url

LynnL4:


grilliiiii (approved):
Thanks for your updates!

PR #1018, opened 2024-03-11
Update: slug AWS edge box is corrected 

KasunThushara:
ADD: Edgebox RPI AWS and Update : AWS and reTerminal DM


grilliiiii (approved):
Thanks for your updates!

PR #1035, opened 2024-03-18
FIX: add 2.13inch eink support into the xiao-eink-expansion-board

peterpanstechland:


grilliiiii (approved):
Thanks for your updates!

PR #1043, opened 2024-03-22
Add: grove vision ai v2 train model picture

limengdu:


grilliiiii (approved):
Thanks for your updates!

PR #1057, opened 2024-03-27
Updated dead Zephyr project links

grahas:
Zephyrs website looks like it got updated, updated links to match


grilliiiii (approved):
Thanks for your updates!

PR #1067, opened 2024-03-29
Add: XIAO C6 get_started page

Love4yzp:


grilliiiii (approved):
Thanks for your updates!

PR #1068, opened 2024-03-30
Fix example code for Grove Air Quality Sensor v1.3

duplys:
The example code gives a compile error (`AirQuality.h` not found). The Grove Air Quality Sensor v1.3 library doesn't have that file either. The example from the library works, though (https://github.com/Seeed-Studio/Grove_Air_quality_Sensor/blob/master/examples/Grove_Air_Quality_Sensor/Grove_Air_Quality_Sensor.ino). Maybe the code in the current example is simply outdated.


MatthewJeffson (approved):
Thank you so much for fixing it up for us!

PR #1072, opened 2024-04-01
Update: Format XIAO C6 Table contents

Love4yzp:


MatthewJeffson (approved):
thanks for the update

PR #1079, opened 2024-04-03
add: add learning course to xiao series

Allenkzl:


grilliiiii (approved):
Thanks for your updates!

PR #1102, opened 2024-04-11
Upload XIAO C6 Arduino Pin Usage

Love4yzp:


grilliiiii (approved):
Thanks for your upates!

PR #1110, opened 2024-04-15
Updated the Jetpack5.1.2 firmware burn procedure for product A608

ZhuYaoHui1998:
Updated the Jetpack5.1.2 firmware burn procedure for product A608


MatthewJeffson (approved):
Thanks for the update!

PR #1112, opened 2024-04-15
Add: xiao esp32s3 train model for sscma

limengdu:


grilliiiii (approved):
Thanks for your updates!

PR #1118, opened 2024-04-16
update: update usage about V2 in Linux

Allenkzl:


grilliiiii (approved):
Thanks for your updates!

PR #1124, opened 2024-04-17
Fixed : Typo on Getting started of NEQTO Engine for Linux on reTerminal

KasunThushara:


grilliiiii (approved):
Thanks for your update!

PR #1120, opened 2024-04-17
Add zephyr documentation around Xiao RP2040

Timo614:
Issue: https://github.com/orgs/Seeed-Studio/projects/6?pane=issue&itemId=57293558

I ended up having to do a bit more with this one as Zephyr didn't have support for the board. I based it off the pico 2040 and I put together board support files and some additional examples for SPI, etc. to confirm everything was working as it was my first zephyr board.

I am still trying to debug why my development laptop (it's rather old) is having issues building the zephyr documentation so I can't test the board file. I did see some html generated which looked good but my machine dies mid build so need to retest once I figure out the issues.

Board support for Xiao RP2040 with Zephyr: https://github.com/zephyrproject-rtos/zephyr/pull/71585

xiao_expansion_oled

PR #1126, opened 2024-04-17
Zephyr: Update location of uf2 file given change to underlying wiki page

Timo614:
Relates to: https://github.com/Seeed-Studio/wiki-documents/pull/1120

Noticed I had kept the location of the uf2 file mentioned as being in the applications folder wrt a copied application but I had modified the examples to avoid the need to copy so this can be modified to point to where the samples are built. Sorry about that.

Should I also add a note about waiting for the Xiao RP2040 to merge in zephyr for the PR there? I have been working on the PR to get it to merge so should be soon but not sure on if they will want further changes.


MatthewJeffson (approved):
Thanks for the update!

MatthewJeffson:
And it is fine, my colleague is going to contact the Zephyr Official and see what is it going here.
No need to add the specific information:D

Timo614:
Sounds good, there were a few issues with the PR related to my formatting and they wanted some documentation modifications. I addressed them today so I think in the next few days pending any other issues it should be good to go but will keep an eye on it.

PR #1128, opened 2024-04-18
Zephyr: Use Expansion shield for examples

Timo614:
Further adjustments to the zephyr documentation. Wasn't aware the Xiao expansion board was already supported as a shield by Zephyr. I've gone ahead and updated the examples to use that where possible. I've also made the associated example repo more generic relying on the xiao_i2c etc. Once this merges I'll cleanup the associated example repo so it removes some duplication there.


MatthewJeffson (approved):
Okay! Thank you for the update! Tim

PR #1134, opened 2024-04-19
Zephyr: ESP32C3 documentation / applications

Timo614:
Re: https://github.com/Seeed-Studio/wiki-documents/issues/1046

I've added initial board application logic here. I tried to see if it were possible to run a wifi server demo using zephyr but it appears they removed the associated code around version 3.1 for a sample there as their logic had broken it during refactoring. Not sure if you can think any other examples that may set this board apart and useful here -- I tested all of the associated ones and added images here to further augment it. I added one around controlling an external LED. The board currently [does not support pwm](https://docs.zephyrproject.org/latest/boards/seeed/xiao_esp32c3/doc/index.html) on the zephyr side.


Timo614:
I've pushed up another PR: #1137 which adds a buzzer and sd card example here. It appears even if the board doesn't declare it supports a feature like pwm it seems possible to still use it (I was able to get PWM working for the Xiao ESP32C3 for the blinky_pwm sample using the buzzer in that PR).

MatthewJeffson:
Wow! Thank you so much! I then will close this branch and check the #1137 only.

PR #1135, opened 2024-04-19
Update sidebar index vision v2

Love4yzp:


MatthewJeffson (approved):
Thanks for the update!

PR #1139, opened 2024-04-23
Fix: xiao esp32s3 D/A11 and D/A12 pin clarification

limengdu:


MatthewJeffson (approved):
Thanks for the update!

PR #1142, opened 2024-04-23
fix: xiao c6 fix a wrong info

Allenkzl:


grilliiiii (approved):
Thanks for your updates!

PR #1145, opened 2024-04-23
add: xiao c6 add tips about switch built-in or external antenna

Allenkzl:


grilliiiii (approved):
thanks for your updates！

PR #1147, opened 2024-04-25
Fix: fix h68k sidebar

chenxixiya:


MatthewJeffson (approved):
Thanks for the update!

PR #1151, opened 2024-04-26
fix: grove v2 fix video autoplay

Allenkzl:


MatthewJeffson (approved):
Thanks for the update!

PR #1161, opened 2024-04-30
Update Sidebars of Edge box and reterminal DM page button for backto pages 

KasunThushara:


MatthewJeffson (approved):
Thanks for update!

PR #1166, opened 2024-05-08
update: V2 HA demo update some images

Allenkzl:


MatthewJeffson (approved):
Thanks for the update

PR #1169, opened 2024-05-09
add: V2 add web camera demo

Allenkzl:


MatthewJeffson (approved):
Thanks for the add up!

PR #1187, opened 2024-05-15
Add: XIAO ESP32 Series and Matter and ESP-IDF wiki

limengdu:


grilliiiii (approved):
Thanks for your updates!

PR #1190, opened 2024-05-16
fix: xiaos3 and grove v2 fix image bugs

Allenkzl:


grilliiiii (approved):
Thanks for your updates!

PR #1191, opened 2024-05-16
update: sscma page update some images and link

Allenkzl:


MatthewJeffson (approved):
Thanks for the update!

PR #1192, opened 2024-05-16
Update: Update Flash Jetpack6.0 for reComputer J401.

ZhuYaoHui1998:
Update Flash Jetpack6.0 for reComputer J401.


MatthewJeffson (approved):
Thanks for the update!

PR #1194, opened 2024-05-17
Add missing dependencies in JetPack flash

lakshanthad:
Added missing dependencies when flashing JetPack to NVIDIA Jetson.


MatthewJeffson (approved):
Thanks for the update!

PR #1215, opened 2024-05-23
fix: kafka_xiao wiki add some content

Allenkzl:


limengdu:
Please resize your photos

PR #1226, opened 2024-05-27
ADD: Edgebox Node red Modbus TCP

KasunThushara:


MatthewJeffson (approved):
Thanks for the add up!

PR #1234, opened 2024-05-29
Update:Edgebox nodered modbus tcp slug changed

KasunThushara:


MatthewJeffson (approved):
Thanks for the update

PR #1252, opened 2024-06-04
update: add power supply detail for reComputer R1000

cure123123:


grilliiiii (approved):
Thanks for your updates!

PR #1259, opened 2024-06-07
upgrade: recomputer r1000 get started and assembly guide

cure123123:


grilliiiii (approved):
Thanks for your updates!

PR #1260, opened 2024-06-07
ADD: Modbus RTU and Edgebox with node red 

KasunThushara:


grilliiiii (approved):
Thanks for your updates!

PR #1261, opened 2024-06-08
Zephyr additions, adjustments to flash command

Timo614:
The Xiao round display shield works well with the xiao nrf52840, esp32s3, and esp32c3. The former had no issues with the SD function from it as well so included an example there. Was unable to get the demo working on the SAMD21 as it didn't have enough memory.

In addition I've switched the logic to rely on the `west flash -r uf2` versus the comments re: moving the uf2 file manually. 

This is unrelated to the task for adding the lvgl 9 support for the round display just saw the shield was added recently in passing so wanted to add it for completeness sake.


MatthewJeffson (approved):
Thank you Tim! Really appreciate it!

PR #1263, opened 2024-06-11
docs(XIAO C3): update nums of analog pins

Love4yzp:


grilliiiii (approved):
Thanks for your updates!

PR #1265, opened 2024-06-11
update a note and 2 illustrations

cure123123:


grilliiiii (approved):
Thanks for your updates!

PR #1266, opened 2024-06-11
YOLOv8 Classification model using custom dataset

feiticeir0:
How to use Jetson OrinNX with YOLOv8 to train a classification model using a custom dataset.


MatthewJeffson (approved):
Thanks for the adding up!

PR #1272, opened 2024-06-13
update pictures:move transparent background

cure123123:


MatthewJeffson (approved):
Thanks for the update!

PR #1284, opened 2024-06-19
Add: Industrial download link，add Allxon jetson getting started example

ZhuYaoHui1998:


MatthewJeffson (approved):
thanks for the update

PR #1290, opened 2024-06-21
update:use ubuntu flash recomputer r1000

LJ-Hao:


grilliiiii (approved):
Thanks for your updates!

PR #1292, opened 2024-06-21
update:r1000 install driver

LJ-Hao:


grilliiiii (approved):
Thanks for your updates!

PR #1299, opened 2024-06-23
Update ESP32-Arduino examples to v3 i2s

lucyknada:
Tested locally, based on migration guide and new docs over at: https://docs.espressif.com/projects/arduino-esp32/en/latest/api/i2s.html

This does not update the save to microSD part yet (record_wav), as that needs separate testing.

### DIFF
--- a/docs/Sensor/SeeedStudio_XIAO/SeeedStudio_XIAO_ESP32S3_Sense/XIAO_ESP32S3_Sense_mic.md
+++ b/docs/Sensor/SeeedStudio_XIAO/SeeedStudio_XIAO_ESP32S3_Sense/XIAO_ESP32S3_Sense_mic.md
@@ -64,6 +64,10 @@ For the first project case, let's detect the noise in the environment and show t
 
 The following is the complete sample program.
 
+:::tip
+Check and make sure which version of ESP32-Arduino you are using, the following example is for 2.0.x the one below is for 3.0.x and later
+:::
+
 ```cpp
 #include <I2S.h>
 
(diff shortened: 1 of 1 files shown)


lucyknada:
tested and porting the wav recording example works similarly too, if this gets merged I can add a PR for that too.

limengdu:
Thank you for submitting your update, our engineers are in the process of verifying it, please wait. @Allen-Kuang

Allenkzl:
Hello  @lucyknada ,  thank you for your commit~ 
I would like to ask you a question. There was a error when I compile your code, which is about the "setPinsPdmRx" method. I attach an image so that you can check if you met this problem before. Another image is about the usage in offical document and it seems different to your usage.
I2S_official
I2S_fix

lucyknada:
@Allen-Kuang are you using v3 of esp32-arduino? these functions are only in 3.0.x onward, v2 still relies on setAllPins()

Allenkzl:
Yess~  My ESP32 version is  3.0.1 @lucyknada 
I2S_fix2

lucyknada:
@Allen-Kuang I tried it with a fresh installation and never used esp32s3-sense; it still works.

image

something in your setup is making it try to resolve for alternatives; none of my setups do, so I'd investigate that; maybe just reinstall the esp32 library or uninstall the other esp32 provider by arduino:

image

and try again; I assume it's providing headers too possibly.

Allenkzl:
In fact, almost all of those using I2S are XIAO ESP32-S3 Sense, so we must arrange time to solve this problem. Thank you for your comtribution again~ I will try the method you mentioned above. If it has any progress I will let you know ~ @lucyknada

limengdu:
image
There's nothing wrong on my end.

limengdu (approved):

Allenkzl:
I use a new enviroment to test it, it works~
Indeed, it's my old enviroment issuse.

PR #1316, opened 2024-06-26
add a picture to retreminal dm get started

cure123123:


grilliiiii (approved):
Thanks for your updates!

PR #1319, opened 2024-06-26
ADD: modbus tcp and reterminal DM with node red

KasunThushara:


grilliiiii (approved):
Thanks for your updates!

PR #1333, opened 2024-07-01
update:reterminal-dm-flash-OS and recomputer_r1000_flash_os

LJ-Hao:


grilliiiii (approved):
Thanks for your updates!

PR #1358, opened 2024-07-04
add:llama-factory

yuyoujiang:
add new wiki for jetson about fine-tune LLM


grilliiiii (approved):
Thanks for your updates!

PR #1381, opened 2024-07-09

nfs0619:
Adjust formatting and enhance content for the WiFi section of the XIAO ESP32S3 (Sense) wireless communication front page in the Chinese wiki documentation.


grilliiiii (approved):
Thanks for your updates!

PR #1382, opened 2024-07-09

nfs0619:
Correct the format of the Bluetooth usage and camera usage sections in the XIAO ESP32S3 (Sense) part of the Chinese Wiki documentation.


MatthewJeffson (approved):
Thanks for the update!

PR #1383, opened 2024-07-10
To supplement and modify the Chinese Wiki

nfs0619:
Make format corrections and add content to the "Microphone Usage" to "Keyword Spotting" sections in the Chinese documentation of XIAO ESP32S3.


MatthewJeffson (approved):
Thanks for the update!

PR #1384, opened 2024-07-10
add firmware link

Jessie219-web:
changed the slug of the respeaker introduction doc, as the link has already been used in the sticker.


MatthewJeffson (approved):
Thanks for the update!

PR #1393, opened 2024-07-12
We have removed the description of flashing Jetpack on J401 using VMware and changed the location of the robotics directory.

ZhuYaoHui1998:
We have removed the description of flashing Jetpack on J401 using VMware and changed the location of the robotics directory.


grilliiiii (approved):
Thanks for your updates!

PR #1397, opened 2024-07-12
fixed a math error and optimized format

Harrison-Xu:


MatthewJeffson (approved):
Thanks for the update!

PR #1416, opened 2024-07-18
Add configure different IP addresses for each POE interface on reServ…

ZhuYaoHui1998:
Add configure different IP addresses for each POE interface on reServer Industrial, Add reactivate the wireless network card driver on reComputer J4012 in Flash Jetpack 6.0


grilliiiii (approved):
Thanks for your updates!

PR #1426, opened 2024-07-19
Update recomputer_R.md, the supported OS

cure123123:


grilliiiii (approved):
Thanks for your updates!

PR #1437, opened 2024-07-22
update latest jp6 download link for reComputer Industrial

yuyoujiang:
update latest jp6 download link for reComputer Industrial


grilliiiii (approved):
Thanks for your updates!

PR #1440, opened 2024-07-23
update：Modify the issues in Wio-pr

nfs0619:
Modify the issues in Wio-pr and include them in the table of contents of the Chinese document.


grilliiiii (approved):
Thanks for your updates!

PR #1487, opened 2024-08-02
update: update supplement for xiao about platform IO

Allenkzl:


grilliiiii (approved):
Thanks for your updates!

PR #1502, opened 2024-08-07
fixed download link of reComputer and reServer BSP packages

ZhuYaoHui1998:
fixed download link of reComputer and reServer BSP packages


grilliiiii (approved):
Thanks for your updates!

PR #1505, opened 2024-08-07
ADD: Home Assistant and reComputer R1000 Modbus

KasunThushara:


grilliiiii (approved):
Thanks for your updates!

PR #1528, opened 2024-08-12
 update link in reComputer industrial J4012 BSP

ZhuYaoHui1998:
update link in reComputer industrial J4012 BSP


MatthewJeffson (approved):
Thanks for the update!

PR #1531, opened 2024-08-12
update link in reComputer another BSP

ZhuYaoHui1998:
update link in reComputer another BSP


MatthewJeffson (approved):
Thanks for the update!

PR #1542, opened 2024-08-13
add:AI NVR demo for reServer J4012

yuyoujiang:
1. add new demo for reServer J4012
2. fix  tar command of flash jetpack to reComputer J4012
3. add hyperlinks for Generative_AI_Intro


grilliiiii (approved):
Thanks for your updates!

PR #1569, opened 2024-08-16
update jetspack download links

yuyoujiang:


MatthewJeffson (approved):
Thanks for the update

PR #1579, opened 2024-08-19
change figure of reCompute J4012

ZhuYaoHui1998:
change figure of reCompute J4012


grilliiiii (approved):
Thanks for your updates!

PR #1581, opened 2024-08-19
add: getting started with j501 carrier board

yuyoujiang:


grilliiiii (approved):
Thanks for your updates!

PR #1600, opened 2024-08-23
add power png to Edgebox-ESP-100-Arduino

ackPeng:


MatthewJeffson (approved):
Thanks for the update!

PR #1602, opened 2024-08-23
update:r1000 assembly guide and yolov8 benchmark

LJ-Hao:
update:r1000 assembly guide and yolov8 benchmark


grilliiiii (approved):
Thanks for your updates!

PR #1605, opened 2024-08-24
Add vlm with speech

ZhuYaoHui1998:
Add vlm with speech


MatthewJeffson (approved):
Thanks for the add up!

PR #1608, opened 2024-08-26
fix type: kicad_sym file link

Love4yzp:
fix type: kicad_sym file link


MatthewJeffson (approved):
thanks for the update!

PR #1618, opened 2024-08-27
update:speech vlm

yuyoujiang:
Update video link


MatthewJeffson (approved):
Thanks for the update

PR #1656, opened 2024-09-08
add: j501 interface usage wiki

yuyoujiang:


nfs0619 (approved):
Thanks for the update

PR #1659, opened 2024-09-09
change link about J1020v2 flash jetpack

ZhuYaoHui1998:
change link about J1020v2 flash jetpack


nfs0619 (approved):
Thanks for the update

PR #1661, opened 2024-09-09
update:yolov8s benchmark

LJ-Hao:
update:yolov8s benchmark


grilliiiii (approved):
Thanks for your updates!

PR #1672, opened 2024-09-13
Update Fuxa installtion reTerminal DM

KasunThushara:


nfs0619 (approved):
Thanks for the update

PR #1675, opened 2024-09-14
delete "you need to run this step after system boot up every time."

ZhuYaoHui1998:
delete "you need to run this step after system boot up every time."


nfs0619 (approved):
Thanks for the update

PR #1686, opened 2024-09-18
fix: added variant ATH20

CorwinDev:
When using both `Grove - VOC and eCO2 Gas Sensor (SGP30)` and `Grove - Temperature and Humidity Sensor (AHT20)`, the temperature wouldn't work, when added this it worked


nfs0619 (approved):
Thanks for the update

PR #1690, opened 2024-09-20
Update screen orientation instructions for reTerminal-DM

gerrygralton:


nfs0619 (approved):
Thanks for the update

PR #1692, opened 2024-09-20
fix resources of reComputer J401 

yuyoujiang:


nfs0619 (approved):
Thanks for the update

PR #1696, opened 2024-09-20
fixed reServer Industrial Jetpack 6 POE enable bug

ZhuYaoHui1998:
fixed reServer Industrial Jetpack 6 POE enable bug


nfs0619 (approved):
Thanks for the update

PR #1697, opened 2024-09-20
add:Gapi on Jetson

yuyoujiang:


nfs0619 (approved):
Thanks for the update

PR #1704, opened 2024-09-23
update Nvidia_Jetson contexts

ZhuYaoHui1998:
update Nvidia_Jetson contexts


nfs0619 (approved):
Thanks for the update

PR #1715, opened 2024-09-26
Update CN-XIAO ESP32S3 Application

nfs0619:


MatthewJeffson (approved):
Thanks for the add up!

PR #1718, opened 2024-09-29
update:add gapi to jetson home page

yuyoujiang:


nfs0619 (approved):
Thanks for the update

PR #1723, opened 2024-09-30
Update: reTerminal and HA under bookworm OS 

KasunThushara:


nfs0619 (approved):
Thanks for the update

PR #1725, opened 2024-09-30
add use nvc on reComputer jetson

ZhuYaoHui1998:
add use nvc on reComputer jetson


grilliiiii (approved):
Thanks for your updates!

PR #1739, opened 2024-10-09
[fix] changed recamera files

litxaohu:
Hello, this is Parker
I changed sidebars.js about recamera, please merge it!


nfs0619 (approved):
Thanks for the update

PR #1743, opened 2024-10-09
fix:reComputer Industrial J30/40

yuyoujiang:


nfs0619 (approved):
Thanks for the update

PR #1768, opened 2024-10-14
ADD: thingsboard and recomputer r1000

KasunThushara:


nfs0619 (approved):
Thanks for the update

PR #1789, opened 2024-10-21
add reComputer R1000 v1.1 Pin Assignment

ackPeng:


nfs0619 (approved):
Thanks for the update

PR #1794, opened 2024-10-22
fix error url for j202 'get one now' and add pcie using about J202 

ZhuYaoHui1998:
fix error url for j202 'get one now' and add pcie using about J202


nfs0619 (approved):
Thanks for the update

PR #1804, opened 2024-10-25
add recomputer r1000 V1.1 description and modify modbus functions description

ackPeng:


nfs0619 (approved):
Thanks for the update

PR #1805, opened 2024-10-25
Add: thingsboard and reComputer r1000 creating DashBoards

KasunThushara:


nfs0619 (approved):
Thanks for the update

PR #1808, opened 2024-10-28
fix some bug about A603 USB ssh

ZhuYaoHui1998:
fix some bug about A603 USB ssh


nfs0619 (approved):
Thanks for the update

PR #1825, opened 2024-10-31
fix wording, typo and punctuation for serial communication

mi-hol:


nfs0619 (approved):
Thanks for the update

PR #1842, opened 2024-11-05
fix: Fixed missing underline in `config.txt` file for R1000 V1.0.

bclswl0827:


nfs0619 (approved):
Thanks for the update

PR #1846, opened 2024-11-06
docs(xiao-rp2040): update code snippet for xiao rp2040

Love4yzp:


nfs0619 (approved):
Thanks for the update

PR #1848, opened 2024-11-06
update: update 2350 article order and name, update watcher to nodered…

Allenkzl:
… interface


limengdu (changes requested):
sidebar不要用一样的序号

PR #1868, opened 2024-11-10
Fix XIAO_BLE.md

esplo:
I found a marge conflict in the starter document.

In this commit, I removed the first explanation. If the other change should be adopted, please provide your comment.


MatthewJeffson (approved):
Thanks for the update and point out!

PR #1878, opened 2024-11-12
change image download linlk

jjjadand:
change reserverJ50 and a603


nfs0619 (approved):
Thanks for the update

PR #1885, opened 2024-11-13
restore slug

jjjadand:
restore slug


nfs0619 (approved):
Thanks for the update

PR #1889, opened 2024-11-14
docs: Add instruction on new solution to RS485 flow control issue

bclswl0827:
This PR added instruction on fixing RS485 flow control issue on reComputer R1000 by injecting kernel module.


grilliiiii (approved):
Thanks for your updates!

PR #1893, opened 2024-11-15
[fix] change recamera files

litxaohu:
Hi this is Parker
I changed recamera files, please merge it!


nfs0619 (approved):
Thanks for the update

PR #1895, opened 2024-11-18
update A603 flashing JP5.1.4 and JP6.1

ZhuYaoHui1998:


nfs0619 (approved):
Thanks for the update

PR #1912, opened 2024-11-20
update xiao esp32c6 pages

Love4yzp:
details see comitt messages.


nfs0619 (approved):
Thanks for the update

PR #1998, opened 2024-12-13
fix hardware image

Carla-Guo:


nfs0619 (approved):
Thanks for the update

PR #2000, opened 2024-12-16
fix: Update driver installation argument for reTerminal DM

bclswl0827:


nfs0619 (approved):
Thanks for the update

PR #2006, opened 2024-12-18
ADD: FTP with rpi 4g LTE hat 

KasunThushara:


nfs0619 (approved):
Thanks for the update

PR #2007, opened 2024-12-18
fix error video of flash jetpack of J401

ZhuYaoHui1998:
fix error video of flash jetpack of J401


nfs0619 (approved):
Thanks for the update

PR #2015, opened 2024-12-19
add: update orin nano devkit

yuyoujiang:


nfs0619 (approved):
Thanks for the update

PR #2021, opened 2024-12-20
Update: 4G LTE hat mqtt python example 

KasunThushara:


nfs0619 (approved):
Thanks for the update

PR #2026, opened 2024-12-20
add minicom configure

LJ-Hao:


nfs0619 (approved):
Thanks for the update

PR #2040, opened 2024-12-24
feat: Added wiki on keyword spotting with TensorFlow Lite

bclswl0827:


nfs0619 (approved):
Thanks for the update

PR #2041, opened 2024-12-25
ADD: 4G LTE hat and mobile hotspot on rpi OS bookworm 

KasunThushara:


nfs0619 (approved):
Thanks for the update

PR #2044, opened 2024-12-25
Add: Setting Up OpenWrt on Raspberry Pi 5 with a 4G LTE Module

KasunThushara:


nfs0619 (approved):
Thanks for the update

PR #2046, opened 2024-12-26
Update: Sub-headings 4g lte hat openwrt

KasunThushara:


nfs0619 (approved):
Thanks for the update

PR #2056, opened 2024-12-30
complete the lerobot projects

ZhuYaoHui1998:
complete the lerobot projects


nfs0619 (approved):
Thanks for the update

PR #2060, opened 2024-12-31
fix some error words spell of A603 Flash Jetpack

ZhuYaoHui1998:


nfs0619 (approved):
Thanks for the update

PR #2066, opened 2025-01-03
update:j501 can usage

yuyoujiang:


nfs0619 (approved):
Thanks for the update

PR #2077, opened 2025-01-07
update: jp6.1 for a605,a607,a6.8

yuyoujiang:


nfs0619 (approved):
Thanks for the update

PR #2084, opened 2025-01-08
docs: Updated the method of controlling peripherals of R1000

bclswl0827:


nfs0619 (approved):
Thanks for the update

PR #2094, opened 2025-01-09
 doc: Added new docs on Frigate NVR

bclswl0827:


nfs0619 (approved):
Thanks for the update

PR #2095, opened 2025-01-09
update the location of raspberrypi files and upload the reComputer_R2…

LJ-Hao:
…000_getting_start


LJ-Hao:
update the location of raspberrypi files

nfs0619 (approved):
Thanks for the update

PR #2101, opened 2025-01-10
add Power consumption and temperature to reComputer_R2000_getting_start

LJ-Hao:


nfs0619 (approved):
Thanks for the update

PR #2105, opened 2025-01-13
update reComputer AI R2130 to AI application

LJ-Hao:


nfs0619 (approved):
Thanks for the update

PR #2114, opened 2025-01-15
Fix: XIAO ESP32C3 Correct the wrong words

ZhangKeLiang0627:


nfs0619 (approved):
Thanks for the update

PR #2124, opened 2025-01-20
Fix: XIAO-ESP32C6 Correct battery usage

ZhangKeLiang0627:
企业微信截图_b7664df6-cc01-41b9-9657-607cbfe6109f


nfs0619 (approved):
Thanks for the update

PR #2136, opened 2025-01-23
Fix: XIAO-nRF52840 Delete an unreasonable line

ZhangKeLiang0627:
895de832bb9d0032a240caa06b64e79c


nfs0619 (approved):
Thanks for the update

PR #2157, opened 2025-02-07
Add flash ubuntu to nvme reComputer_R2000

LJ-Hao:


nfs0619 (approved):
Thanks for the update

PR #2170, opened 2025-02-11
update: jp6.2 of a603 and a608 

yuyoujiang:


nfs0619 (approved):
Thanks for the update

PR #2176, opened 2025-02-12
Update: 4G LTE CAT rpi hat 

KasunThushara:


nfs0619 (approved):
Thanks for the update

PR #2177, opened 2025-02-12
upload MR60 radar GUI software

discodyer:


grilliiiii (approved):
thanks for your updates!

PR #2190, opened 2025-02-14
Update: recomputer r1000 collection page vision sub topic 

KasunThushara:


nfs0619 (approved):
Thanks for the update

PR #2196, opened 2025-02-17
Update: WM1302 Module with new Linux kernel

tonyleo723:


nfs0619 (approved):
Thanks for the update

PR #2199, opened 2025-02-17
edit the wiki-"Hardware Interface" in reComputer Mini

jjjadand:
edit the wiki-"Hardware Interface" in reComputer Mini


nfs0619 (approved):
Thanks for the update

PR #2207, opened 2025-02-18
edit the wiki about reCamera-“Convert and Quantize AI Models”

jjjadand:
edit the wiki about reCamera-“Convert and Quantize AI Models”


nfs0619 (approved):
Thanks for the udpate

PR #2212, opened 2025-02-19
edit reComputer Mini Wiki

jjjadand:
change a png.  change the link of micro-usb about flashing jetpack


nfs0619 (approved):
Thanks for the update

PR #2216, opened 2025-02-19
edit reserver industrial WIKI

jjjadand:
divid the previous WIKI into two new WIKI


nfs0619 (approved):
Thanks for the update

PR #2224, opened 2025-02-22
spanish translation xiao series

MDG336:


nfs0619 (approved):
Thanks for the update

PR #2242, opened 2025-02-27
ADD recomputer r1100 getting stated assemble guide and configuring 

KasunThushara:


MatthewJeffson (approved):
Thanks for the add upate!

PR #2244, opened 2025-02-28
Full Raspberry Pi Product Lineup Spanish Translation w/Name Format Corrections

MrCrabioto:


ghost (commented):
1

nfs0619 (commented):
Hello, thanks for your contribution! I noticed a few issues in this PR:

One issue (see Screenshot 1) is that some documents do not follow the required naming convention for filenames and slugs.
企业微信截图_17410550665547

Another issue (see Screenshot 2) is that there are many English documents included in this PR. Could you check if there was an issue during submission?
企业微信截图_17410551789821

Additionally (see Screenshot 3), it looks like some HTML code in these pages may have been accidentally modified, which is causing errors in the PR.
企业微信截图_17410562138466

Could you review these points and update the PR accordingly? Let me know if you need any help. Thanks! 😊

nfs0619:
Hello again!

Thanks for your updates! I see that you've made progress in fixing the issues, which is great. However, I noticed that some of the original concerns still remain:  

1. **Filename and Slug Issues** – While some documents have been corrected, there are still a few that don’t follow the required naming convention. For reference, the correct format should be:  
   - **Slug format**: `/es/` + the English document's slug.  
   - **File naming**: Should follow this standardized format → `es_OriginalFileName`.  
   See the screenshot for example. There might be more cases beyond this example, so please double-check all affected files. 

企业微信截图_17423472986809

2. **English Documents Still Present** – You've removed some, but there are still a significant number of English documents in this PR. Please review and ensure that only translated content is included. See the example screenshot. There might be more cases beyond this example, so please double-check all affected files.

企业微信截图_17423470032645
企业微信截图_17423472545403

3. **Remaining HTML Code Issue** – Most errors have been fixed, but there is still one problematic file: `/es/edge_box_rpi_200_node_red_modbus_tcp/`. Please check the screenshot for details.  

企业微信截图_17423499177332

Could you take another look and update the PR accordingly? Let me know if you need any help! Thanks again for your contributions! 😊

nfs0619 (approved):
Thanks for the update

PR #2246, opened 2025-02-28
update two-Channel-CAN-BUS-FD-Shield-for-Raspberry-Pi

LJ-Hao:


ghost (approved):
Thanks for the update

PR #2247, opened 2025-02-28
update: H01 and A603

yuyoujiang:


ghost (approved):
Thanks for the update

Issue #2058, opened 2024-12-30
A603 carrier board USB driver problem


boppabu:
Hi,
trying to flash Jetpack on A603 Carrier Board with Jetson Orin Nano 8 GB devkit.
Until now, tried all possible versions and exactly described as the tutorial, was hoping that the issue is fixed with 6.1 but that is still not the case.
https://wiki.seeedstudio.com/reComputer_A603_Flash_System/
Also followed this post here: https://forum.seeedstudio.com/t/a603-carrier-board-issues-flashing-and-usb/276785/37

I am sure that the driver files are replaced properly in  Linux_for_Tegra/ folder, used sudo.
Jetpack is flashing correctly, HDMI works, it boots and I see desktop but can't use a keyboard and mouse as USB is not working.
Host machine is ubuntu 22.04

Can you please support? 
Thanks in advance.

nfs0619:
Thank you for your feedback. We have noted the issue you're experiencing, and the development team is actively investigating it. They will reach out to you with an update once progress is made. We appreciate your patience and understanding!

matt-nolan11:
Hi there, I'm having what sounds like the same exact issue, and it seems like many other are as well. Any progress on this?

neoscell:
Same here

nfs0619:
Hello, we sincerely apologize for the inconvenience caused!  
We are aware of the USB driver issue, and our development team is currently investigating it with urgency to find a solution as soon as possible. To help us better diagnose the problem, could you kindly provide the following information if possible?

1. **JetPack Version**: The specific JetPack version you are using (e.g., JetPack 5.1.2, 6.0, etc.).  
2. **Power Supply Information**: The specifications of the power supply you are using (voltage, current, etc.).  
3. **SSD Model**: The model of the SSD you are using.  
4. **Operation Logs**: If you could provide your relevant operation logs, it would greatly assist in our analysis.

Thank you again for your patience and support! If you have any further findings or information, please feel free to let us know, and we will continue to follow up.

neoscell:
Jetson Orin Nano 8 GB
Ubuntu 22.04.5

1) Attempted with JetPack 6.0 & 6.1

2) Power Supply from Seed Studio - GST90A19-P1M - 19V / 4.75A / 90W Max

3) GamerKing 256 GB SSD 

4) Code used fo flashing: 
```
tar xf Jetson_Linux_R36.4.0_aarch64.tbz2
[...]

PR #2252, opened 2025-03-03
Update A603 Carrier Board JP62 Docs with some clarity about the driver installation pre flash

johnmckenna-snd:
I had a problem where I didn't notice the tar command was failing for the USB drivers because the download wasn't in the right place. I subsequently flashed the Jetson without the USB drivers and you'll be surprised to find out that the USB ports didn't work. After much banging my head against various walls, I found out I messed up and didn't inject the drivers. I did and the USB ports worked. Woot.

I have created this PR to add a blurb about this in the docs to hopefully help others from going down this dark and twisted path.

[Issue Here](https://github.com/Seeed-Studio/wiki-documents/issues/2058)

### DIFF
--- a/docs/Edge/NVIDIA_Jetson/Carrier_Boards/A603/A603_Flash_JetPack.md
+++ b/docs/Edge/NVIDIA_Jetson/Carrier_Boards/A603/A603_Flash_JetPack.md
@@ -532,11 +532,9 @@ wget https://developer.nvidia.com/downloads/embedded/l4t/r36_release_v4.3/releas
 
 **Step 2:** Assemble the Flashing Package
 
-Please note you will likely need to move the USB drivers you downloaded into the directory you have downloaded the Jetson Linux tools. For example if you opened your terminal and didn't change directories, you are likely in your user home `~/` folder, and if you downloaded the drivers via the browser, the drivers are likely in `~/Downloads`.
+Please note that we need to put the Nvidia driver and the peripheral drivers in the same directory, and then open the terminal in that directory and execute the following code:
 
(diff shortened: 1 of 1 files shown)


yuyoujiang:
Hi johnmckenna-snd, thank you very much for your suggestion. I'm really excited about your pull request.
I think it would have a better effect if we add a picture here. And we can describe it like this:

----------------------------
...
Please note that we need to put the Nvidia driver and the peripheral drivers in the same directory, and then open the terminal in that directory and execute the following code:

drivers

```bash
tar xf Jetson_Linux_r36.4.3_aarch64.tbz2
sudo tar xpf Tegra_Linux_Sample-Root-Filesystem_r36.4.3_aarch64.tbz2 -C Linux_for_Tegra/rootfs/
sudo tar zxpf 603_jp62.tar.gz
sudo cp -r 603_jp62/Linux_for_Tegra/* Linux_for_Tegra/
cd Linux_for_Tegra/
sudo ./tools/l4t_flash_prerequisites.sh
sudo ./apply_binaries.sh
```
...
---------------------------
If you think this approach is feasible, please modify the PR.

johnmckenna-snd:
Sure.

I am unsure how to add that image specifically. It seems like most of your images are hosted at `files.seeedstudio.com`. Does this image have a URL?

nfs0619:
@johnmckenna-snd Thanks for your help!

You can use this URL for the image:
https://files.seeedstudio.com/wiki/reComputer-Jetson/A603/driver_files_directory_layout.png

Let me know if you need anything else!

nfs0619 (approved):
Thanks for the update

PR #2256, opened 2025-03-04
NVIDIA Spanish file names standarization

OCR1A:
I have updated the Nvidia directory to es_NameOfFileInSpanish spanish standarized file names.

Regards.


ghost (approved):
Thanks for the update

PR #2270, opened 2025-03-07
update: Update bazzar link for 6ch relay

discodyer:


nfs0619 (approved):
Thanks for the update

PR #2275, opened 2025-03-10
docs: Additional notes on the use of DI/DO ports for R1100

bclswl0827:


nfs0619 (approved):
Thanks for the update

PR #2280, opened 2025-03-11
EOL: xiao nrf52840 mic EI wiki

limengdu:


nfs0619 (approved):
Thanks for the update

PR #2290, opened 2025-03-12
chore: Make Frigate tutorial work again

bclswl0827:


nfs0619 (approved):
Thanks for the update

PR #2294, opened 2025-03-13
Add xiao rp2350 nuttx

rcsim:
This PR adds the NuttX Page support for XIAO-RP2350. Also move NuttX Page for XIAO-RP2040 to same level as other RTOSes (Zephyr).


nfs0619 (approved):
Thanks for the update

PR #2298, opened 2025-03-13
update pi-gen-expand to raspberry pi device

LJ-Hao:


nfs0619 (approved):
Thanks for the update

PR #2324, opened 2025-03-20
Add: Pi 4G LTE HAT QMI mode 

KasunThushara:


nfs0619 (approved):
Thanks for the update

PR #2343, opened 2025-03-26
add reComputer Super Getting started

ZhuYaoHui1998:
add reComputer Super Getting started


nfs0619 (approved):
Thanks for the update

PR #2354, opened 2025-03-28
update recamera doc

cure123123:


nfs0619 (approved):
Thanks for the update

PR #2361, opened 2025-03-31
update Raspberry_Pi_Devices  wiki

LJ-Hao:


nfs0619 (approved):
Thanks for the update

PR #2366, opened 2025-04-01
add: new bsp of jetson

yuyoujiang:


nfs0619 (approved):
Thanks for the update

PR #2367, opened 2025-04-01
Reorganize the hardware interface section of the reCamera Gimbal and add detailsReorganize the hardware interface section of the reCamera Gimbal and add details

litxaohu:
Please merge it!


nfs0619 (approved):
Thanks for the update

PR #2369, opened 2025-04-02
Update Replacement_LCD_Screen_for_DSO_nano.md

mouseart:
Correction error: The 1 × 3.5MM TSR connector should be the 1 × 3.5MM TRS connector.


nfs0619 (approved):
Thanks for the update

PR #2426, opened 2025-04-14
Xiao c3 iBeacon wiki page added

Priyanshu0901:
#2422 
This PR adds a low-power temperature sensing system using the Seeed Studio XIAO ESP32-C3.   
It broadcasts temperature data via BLE in iBeacon format and supports approx.


MatthewJeffson:
Wow! @Priyanshu0901 Just in timeXD

nfs0619 (approved):
Thanks for the update

PR #2471, opened 2025-04-26
edit Orin flash, add link

NU-iz-da-fam:
I update link to check Host PC OS. 

And also update table with L4T download link for feasibility.


nfs0619 (approved):
Thanks for the update

PR #2490, opened 2025-05-02
Added default username & password and alternate setup default IP address

HardwareHarry:
Added the default username and password (the forced change password prompt on a new setup covers up the normal login page where the defaults are shown.  I couldn't find the default username and password in any documentation.

I also provided the alternate IP address of 192.168.16.1 which is used for WiFi AP mode setup, which wasn't referenced here.


nfs0619 (approved):
Thanks for the update

PR #2563, opened 2025-05-26
Update A608_Flash_JetPack.md

EdwardNickson:
Fix filenames in J6.2 step 3


nfs0619 (approved):
Thanks for you update!

PR #2717, opened 2025-07-04
Update XIAO_ESP32S3_Pin_Multiplexing.md

saurabhchakrabarti:
### Description

This pull request fixes the example code for configuring the second hardware serial port (`MySerial1`) on the Seeed Studio XIAO ESP32S3.

#### Changes Made:
- Corrected the `Serial.begin()` statement to match the correct order of parameters: `(baud rate, config, RX pin, TX pin)`.
- Updated the example line to:

  ```cpp
  MySerial1.begin(115200, SERIAL_8N1, D9, D10);

This ensures the serial port is correctly initialized using the intended RX (D9) and TX (D10) pins with a baud rate of 115200 and 8N1 configuration.


nfs0619 (approved):
Thanks for the update

PR #2991, opened 2025-09-05
Fixed some issues in the ESPhome firmware of MR60BHA2

lewuq:
1. Removed the content that would lead to failure.
2. Added some contents related to Wi-Fi and OTA to avoid Wi-Fi connection failure.
3. Added some contents related to API.
4. Added some comments to aid understanding.
5. Removed the "rmt_channel" content from the light section, as it is not present in the library.
6. Used the recommended content from the ESPhome documentation for binary_sensor and sensor sections, excluding "fall_detect".


limengdu (changes requested):
代码版本使用3.5再跑一遍，非必要不要删除代码

PR #3431, opened 2025-11-05
I added the project code to the wiki by cloning the repository with git clone.

kouroshkarimi:
add the code by cloning it when the code needs to be run in the wiki.


nfs0619 (approved):

nfs0619 (approved):
Thank you for the update!

PR #3833, opened 2026-01-03
Fix slug for XIAO ESP32C6 pin multiplexing

nopeless:


nopeless:
I made two PRs by accident but this is a completely separate typo 🙏

nfs0619 (approved):
Thanks for the update!

PR #3882, opened 2026-01-08
Include ESPHome configuration in getting started guide

berikv:
Added ESPHome template for XIAO ESP32C6.


limengdu:
Thank you for your contribution to the documentation. In accordance with Seeed's documentation guidelines, the content is not well-suited for the Getting Started category. We will consider creating a dedicated Home Assistant section to cover related tutorials. We appreciate your understanding and cooperation.

berikv:
@limengdu I'm looking for a place to document the behaviour of the antenna selector. Without adding specific code, the antenna selector is disabled, and WiFi connections are hard to establish. Adding a minimal example to the ESP32C6 like the one in this PR would make it much more obvious.

Only providing Android examples is dubious in my opinion. Why not split up the example section and allow different tech stacks their own page?

berikv:
FYI, I opened a PR for PlatformIO for the same reason: https://github.com/platformio/platformio-docs/pull/395/commits/cf59551e775a3ff9c6e5598088c6a027b8db424c
No action from their side yet.
And added the antenna selector pin definition to ESPHome, which was merged and released in v2026.1.1: https://github.com/esphome/esphome/pull/13480

For more info, see this discussion: https://forum.seeedstudio.com/t/xiao-esp32c6-switching-between-builtin-and-external-antenna/276374

PR #3920, opened 2026-01-14
fix: Escape tilde to avoid rendering as strike-through text

jamesharr:
The tilde in the docs render as strikethrough, this is a small fix

Illustration of problem:
image

Link to page:
- https://wiki.seeedstudio.com/recomputer_r/#:~:text=Supply%20Voltage%28AC%2FDC


nfs0619 (approved):
Thank you for your correction!

PR #3989, opened 2026-01-22
Change ESP32 framework type from Arduino to ESP-IDF

paull27:
esp-idf is the [default framework](https://github.com/esphome/esphome/pull/12746) for ESP32 in esphome 2026.1. There is no need to use the arduino framework specifically.


Allenkzl:
Thank you very much for your PR~
I will leave a comment over there and say user can switch to esp-idf if they want. Because I worry about some function incompatiable if we switch to esp-idf directly.

PR #4096, opened 2026-02-09
docs: update reComputer industrial docs and add R22xx series

xinjiechen727:
update reComputer industrial docs and add R22xx series and updated the descriptions related to R21/R22.


nfs0619 (changes requested):
不要更改 docusaurus 的版本

PR #4498, opened 2026-04-20
Added I2S vs USB firmware information for reSpeaker XVF3800 with XIAO

HerrB92:
Hi there,

as my newly delivered reSpeaker XVF3800 with XIAO was in I2S mode I expect that this is the default. This may create frustration, as the device may seem to work (flashing LEDs etc.), but is actually not visible/detected as USB device in Windows and Linux.

I have therefore added a section on getting started with I2S and USB firmware for the reSpeaker XVF3800, default I2S mode and visibility as USB device as well as using safe mode to switch.

Best Regards,

HerrB92


Frank-Nie02 (approved):
Thanks for your contribution and for improving the documentation. This is really helpful!

PR #4608, opened 2026-05-11
Submit the low-power usage steps for XIAO RP2350

lewuq:
- modified the content of the pin map and changed SPIO to SPI0.
- added steps for how to use low-power mode.


limengdu (approved):

PR #4625, opened 2026-05-14
fix ESP32-S3 SD card data output pin

Lukilyy:


limengdu (approved):

PR #4696, opened 2026-05-27
Update: XIAO-RP2350-with-Arduino.md - simplify Flash the firmware sec…

Carla-Guo:
Simplify the Flash the firmware section in XIAO RP2350 Arduino wiki by removing Arduino code and explanation, keeping only the UF2 firmware flashing method. Added notes about the third-party firmware source (peterharperuk/pico-examples) and pico-sdk dependency with compilation warnings.


limengdu (approved):

PR #4796, opened 2026-06-10
Fix typo

mrdarrengriffin:
Fixes typo in heading. "Hardware Oview" to "Hardware Overview"


nfs0619 (approved):
Thanks for catching and fixing that typo. Much appreciated!

PR #4894, opened 2026-06-22
Include 1-Wire sensor example for DS18B20

WeSpeakEnglish:
Added example code for Dallas/Maxim DS18B20 sensor using 1-Wire protocol.

### DIFF
--- a/sites/en/docs/Sensor/SeeedStudio_XIAO/SeeedStudio_XIAO_SAMD21/Seeeduino-XIAO.md
+++ b/sites/en/docs/Sensor/SeeedStudio_XIAO/SeeedStudio_XIAO_SAMD21/Seeeduino-XIAO.md
@@ -355,6 +355,108 @@ void loop()
 
 <p style={{textAlign: 'center'}}><img src="https://files.seeedstudio.com/wiki/Seeeduino-XIAO/img/THonny_DAC.jpg" alt="pir" width={600} height="auto" /></p>
 
+#### 1-Wire
+
+In this example Dallas/Maxim DS18B20 sensor connected to the line DA2.
+
+```cpp
+#include <Arduino.h>
+
+#define OW_PIN 2
+
+// Bus control
+void    ow_low()  { pinMode(OW_PIN, OUTPUT); digitalWrite(OW_PIN, LOW); }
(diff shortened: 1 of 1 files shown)


limengdu:
Hi @WeSpeakEnglish, thank you for contributing this 1-Wire DS18B20 example! The code is clean and well-structured. Here are a few suggestions to make it even better for wiki readers:

### 1. Add wiring instructions
The example jumps straight into code without explaining the hardware setup. DS18B20 requires a **4.7kΩ pull-up resistor** between the data line and VCC. Could you add a brief wiring description, e.g.:

- DS18B20 VCC → 3.3V
- DS18B20 GND → GND  
- DS18B20 DQ → XIAO pin A2 (DA2)
- 4.7kΩ resistor between DQ and VCC

### 2. Clarify the pin mapping
The code uses `#define OW_PIN 2`, which maps to A2/DA2 on the SAMD21. It would help to mention this explicitly in the text above the code block so users don't have to guess.

### 3. Parasitic power note
The code uses parasitic power mode (`digitalWrite(OW_PIN, HIGH)` during conversion). This is fine but worth mentioning — some users may prefer normal 3-wire power, and parasitic mode can be unreliable with long cables or multiple sensors.

### 4. Multi-device limitation
The example uses Read ROM (`0x33`) which only works with a single device on the bus. A short note like "*This example assumes a single DS18B20 on the bus. For multiple devices, use Search ROM (`0xF0`) instead*" would prevent confusion.

### 5. Suggested library alternatives
For users who want a quicker start, you could mention existing Arduino libraries:
- [OneWire](https://github.com/PaulStoffregen/OneWire) + [DallasTemperature](https://github.com/milesburton/Arduino-Temperature-Control-Library) — the standard approach

These are all optional improvements. The code itself works correctly — nice CRC implementation and clean serial output. Let us know if you'd like to address any of these or if you prefer to merge as-is!

nfs0619:
Hi @WeSpeakEnglish , thank you again for your contribution.

Since there hasn’t been any update for some time, we’ll close this PR for now. The example is appreciated, but the current version still needs a few adjustments to better fit our wiki/tutorial style.

If you’d still like to continue with this contribution, please feel free to reopen this PR or submit a new one after addressing the previous suggestions. We’d be happy to review it again.

Thanks again!

PR #5164, opened 2026-07-22
add E1003 3D model STP file

Lukilyy:


Lukilyy (commented):
add E1003 3D model STP file

PR #5336, opened 2026-08-07
docs: add FAQ for Arducam OV9281 (Jetvariety) on the A603 carrier board

vdmaas98:
Adds a new FAQ page: Use Arducam OV9281 (Jetvariety) Camera with A603 Jetson Carrier Board, registered in the sidebar next to the existing IMX477-on-A603 FAQ.

Arducam Jetvariety cameras do not work on the A603 with Arducam's own installer, because its overlay targets the developer-kit wiring (cam_i2cmux, serial_b/serial_c) while the A603's single CSI port is serial_a on i2c-2. There is currently no wiki page covering this combination, and getting it working involves one dangerous trap: booting the installer's dev-kit boot entry disables all USB on the A603, which together with the keyboard-less boot menu means a full lockout.

The page documents the working configuration (Arducam kernel plus a sensor-swapped A603 device tree), the boot safety procedure, a ribbon-cable diagnostic, and working v4l2 capture settings. Everything was verified on real hardware: Jetson Orin Nano 8GB on an A603, JetPack 6.2 (L4T R36.4.3), OV9281 B0223, August 2026.

Changes:
- sites/en/docs/Edge/NVIDIA_Jetson/FAQs/Use_Arducam_OV9281_Camera_with_A603.md (new page)
- sites/en/sidebars.js (one entry, next to Use_IMX477_Camera_with_A603)

PR #5337, opened 2026-08-07
docs: add FAQ for fully headless A603 setup, internet sharing and boot recovery

### LINKED ISSUE #5213: Headless upgrades of A603 is very painful
Thanks so much for the A603 carrier - it's a very promising hardware product!

Unfortunately there are significant practical challenges with it using it beyond an experimental platform. Because the carrier doesn't integrate with Nvidia's SDKManager we have to use the SeedStudio process which is very manual, and error prone, cut/and/paste (and there are documentation issues - for example, the JP 7.2 instructions say to use USB C to flash the board - which is not possible, you have to use the USB micro connector).

The biggest issue is once the initial flash is done, there is no way to provision the system over USB anymore - you have to connect a screen and keyboard and manually provision connectivity to continue. Not only is this inconvenient if the carrier is part of a larger system you have to arrange for a screen and keyboard to always be near by to upgrade it, which is not necessary for a regular Nvidia carrier which just needs USB and a jumper.

I understand that integrating with the SDKManager may be very challenging, but at the very least allowing configuration of connectivity headless would be a big step forward. Also the links to sharepoint for drivers are quite problematic since they can't be retrieved with just a "wget", so just developing some simple shell scripts to make the install less painful is challenging.

Any improvements you could make to make upgrading more convenient would make the product much more useful.

Thanks,

vdmaas98:
Adds a new FAQ page: Headless Setup, Internet Sharing and Boot Recovery for the A603, registered in the sidebar.

Motivated by issue #5213 (Headless upgrades of A603 is very painful). The A603 can in fact be provisioned and operated entirely over the W6 Micro-USB gadget network, with no monitor, keyboard, Ethernet or WiFi at any point, but none of this is currently documented: the gadget link at 192.168.55.1, the charge-only-cable trap, host-side NAT for giving the board internet (verified macOS pf ruleset with its ordering gotchas, plus the Linux equivalent), the apt IPv6 hang and its IPv4 pin, and, most importantly, how to survive boot-configuration experiments on a board whose boot menu accepts no keyboard input: an auto-revert systemd timer pattern, and last-resort RCM recovery through the flashing initrd without flashing (W7 pin 3 to 4, no recovery button).

Everything on the page was used in practice on an A603 with a Jetson Orin Nano 8GB on JetPack 6.2 (L4T R36.4.3), operated headless from day one.

Changes:
- sites/en/docs/Edge/NVIDIA_Jetson/FAQs/Headless_Setup_and_Recovery_for_A603.md (new page)
- sites/en/sidebars.js (one entry)

Note: #5336 also touches the FAQ block of sites/en/sidebars.js; whichever merges second may need a trivial rebase.


nfs0619 (approved):
Thank you for your contribution! We really appreciate your help in improving our documentation.

